PR #30, opened 2019-09-05
Persistent settings for view reset

pshriwise:
This PR addresses #28.

When the default view is requested, only the geometric properties of the default view are updated. All settings (highlighting, masking, colors, etc.) from the active view will persist.

An option to reset colors now resides in the color dialog box, which will adopt the cell and material information of the default view (including their colors).

Other unrelated changes:

  - removed a deprecated call to `screenGeometry`
  - resolved a problem with retrieving the `colorDialog/Visible` property for the application. When that property was missing, `None` is returned and the conversion to int would fail. Given that the setting is stored as an integer, this conversion wasn't needed and could be removed. The missing application setting value will now evaluate to `False` and hide the color dialog - which is appropriate.

### DIFF
--- a/openmc-plotter
+++ b/openmc-plotter
@@ -795,10 +795,8 @@ class MainWindow(QMainWindow):
                                                QtCore.QSize(400, 500)))
         self.colorDialog.move(settings.value("colorDialog/Position",
                                              QtCore.QPoint(600, 200)))
-
         is_visible = settings.value("colorDialog/Visible")
         is_visible = bool(is_visible)
-
         self.colorDialog.setVisible(is_visible)
 
     def restoreModelSettings(self):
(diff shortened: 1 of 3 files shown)


pshriwise:
After some digging I got the color options dialog window sorted out. `PySide2` released `5.13.1` yesterday, which possibly has a bug in the `QSettings.value`/`QSettings.setValue` methods.  Either `QSettings.setValue` doesn't recognize the int `0` as a valid settings entry or `QSettings.value` doesn't respect the return of a `0` default value (it seems to return other default values/objects just fine). Or it's both of these things. I'm working on where to report this...

For now, I've modified this setting reading s.t. when `None` is returned from the method we treat it as `False` or `0`. Setting behaviors based on a version check of `PySide2` might be a more official way to handle it, but I think this will do.

PR #49, opened 2020-08-12
remove is_pyqt5()

AI-Pranto:
QtCore.qVersion() instead of is_pyqt5 function

### DIFF
--- a/openmc_plotter/plotgui.py
+++ b/openmc_plotter/plotgui.py
@@ -19,10 +19,7 @@ from .plotmodel import _NOT_FOUND, _VOID_REGION, _OVERLAP, _MODEL_PROPERTIES
 from .scientific_spin_box import ScientificDoubleSpinBox
 from .custom_widgets import HorizontalLine
 
-if QtCore.qVersion():
-    from matplotlib.backends.backend_qt5agg import FigureCanvas
-else:
-    from matplotlib.backends.backend_qt5agg import FigureCanvas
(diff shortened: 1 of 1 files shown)


pshriwise:
Can you provide some motivation for this change? Was there a problem with the way it was before?

AI-Pranto:
Not a big issue, just want to remove the warning message. Is it ok @pshriwise?
```
The is_pyqt5 function was deprecated in Matplotlib 3.3 and will be removed two minor releases later. Use QtCore.qVersion() instead.
  if is_pyqt5():
```

pshriwise:
Ah, yeah that definitely makes sense. I hadn't seen this in matplotlib yet.

openmc_plotter/plotgui.py
```diff
@@ -19,7 +19,7 @@
 from .scientific_spin_box import ScientificDoubleSpinBox
 from .custom_widgets import HorizontalLine
 
-if is_pyqt5():
+if QtCore.qVersion():
     from matplotlib.backends.backend_qt5agg import FigureCanvas
 else:
     from matplotlib.backends.backend_qt5agg import FigureCanvas
```
pshriwise, line 25: Looking at this again, we can probably just remove this check. The lines in under the `if` and `else` are exactly the same...

```suggestion
from matplotlib.backends.backend_qt5agg import FigureCanvas
```

pshriwise:
Thanks @AI-Pranto!

PR #53, opened 2020-12-17
:scroll: added MIT LICENSE

Shimwell:
Adding the same license as OpenMC uses. This will need to be included in the PyPi tar ball when making a conda forge install


pshriwise:
A good start for sure. Thanks @Shimwell!

PR #60, opened 2021-04-13
GHA Updates

pshriwise:
This adds a CI file to run the single test we have on CI for PRs into `develop` and `master`. 

It also updates one of the PyPI scripts to no longer trigger on the `develop` branch. PyPI requires a new version each time it is run, so it doesn't make sense to run it for every PR merge. If there's a better way to handle that @Shimwell, let us know!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,42 @@
+name: CI
+
+on:
+  # allows us to run workflows manually
+  workflow_dispatch:
+
+  pull_request:
+    branches:
+      - develop
+      - main
+  push:
+    branches:
+      - develop
+      - main
+
+env:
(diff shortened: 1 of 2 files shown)


Shimwell:
Looks great, extra nice to see those OpenMC dockerhub images being used.

I noticed you are using openmc/openmc:v0.12.1 which is the latest release. Wondering if you want to tag the latest release version instead so it is pointing to the newest version automatically.

pshriwise:
Yeah I suppose that for development we should make sure we're compatible with the development version of OpenMC. If I'm understanding correctly that is.

paulromano (approved):
This is great! Thanks @pshriwise. As @Shimwell said, really cool to see how easy it is to take advantage of the Docker image.

PR #87, opened 2022-04-26
Adding support for viewing distribcell tallies

pshriwise:
@paulromano must have had the same thought that I did in the NEA workshop today and created #86. Plotting these from the Python API is possible, but tricky. This adds that capability to the plotter.

### DIFF
--- a/openmc_plotter/plotmodel.py
+++ b/openmc_plotter/plotmodel.py
@@ -282,11 +282,17 @@ class PlotModel():
 
     def create_tally_image(self, view=None):
         """
+        Parameters
+        ----------
+        view :
+            View used to set bounds of the tally data
+
         Returns
         -------
-        tuple : image data (numpy.ndarray), data extents (optional),
-                data_min_value (float), data_max_value (float),
-                data label (str)
+        tuple
+            image data (numpy.ndarray), data extents (optional),
+            data_min_value (float), data_max_value (float),
+            data label (str)
         """
         if view is None:
             view = self.currentView
@@ -387,13 +393,11 @@ class PlotModel():
                 mean_data = self._create_tally_domain_image(tally,
                                                             'mean',
                                                             scores,
-                                                            nuclides,
-                                                            view)
+                                                            nuclides)
                 std_dev_data = self._create_tally_domain_image(tally,
                                                            'std_dev',
                                                            scores,
-                                                           nuclides,
-                                                           view)
+                                                           nuclides)
                 image_data = 100 * np.divide(std_dev_data[0],
                                              mean_data[0],
                                              out=np.zeros_like(mean_data[0]),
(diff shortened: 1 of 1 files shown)


pshriwise:
image

pshriwise:
This is still a little rough, so I'm marking it as a draft for now. Just wanted to get it opened while the whole problem was in my head.

openmc_plotter/plotmodel.py
```diff
@@ -467,6 +501,31 @@ def _do_op(array, tally_value, ax=0):
...
+        data = data.flatten()
+
+        cell_id = dfilter.bins[0]
+        # create a mask for ids that match the cell
+        image_data = np.zeros_like(self.ids)
+
+        cell_id_mask = self.cell_ids == cell_id
+        for i, v in enumerate(data):
+            instance_mask = self.instances == i
+            image_data[np.logical_and(cell_id_mask, instance_mask)] = v
```
paulromano:
```suggestion
            image_data[cell_id_mask & instance_mask] = v
```

openmc_plotter/plotmodel.py
```diff
@@ -467,10 +504,32 @@ def _do_op(array, tally_value, ax=0):
 
         return image_data, None, data_min, data_max
 
+    def _create_distribcell_image(self, tally, tally_value, scores, nuclides):
```
paulromano, line 504: Either this needs `view` as a last argument, or the calls to this method should not pass `view`

openmc_plotter/plotmodel.py
```diff
@@ -467,10 +504,32 @@ def _do_op(array, tally_value, ax=0):
...
+        sp = self.statepoint
```
paulromano:
```suggestion
```

openmc_plotter/plotmodel.py
```diff
@@ -467,10 +504,32 @@ def _do_op(array, tally_value, ax=0):
...
+        dfilter = tally.find_filter(openmc.DistribcellFilter)
+
+        data = tally.get_values(scores=scores, nuclides=nuclides, value=tally_value)
+        data = data.flatten()
+
+        cell_id = dfilter.bins[0]
+        # create a mask for ids that match the cell
+        image_data = np.zeros_like(self.ids)
```
paulromano: I tried out this branch and it wasn't working for me; eventually traced it down to this!
```suggestion
        image_data = np.zeros_like(self.ids, dtype=float)
```

Also, for a distribcell tally, should we not set the initial state to all `np.nan` instead of all zeros? It doesn't really make sense to me to assign zero values to locations in the image data where we don't have matching bins.

paulromano (changes requested):

paulromano (approved):
Works great now -- thanks for this awesome feature!

PR #93, opened 2022-05-04
Save "Enable Overlap Coloring" settings in edit tab

kkiesling:
This saves the "Enable Overlap Coloring" setting state in the edit menu when the plotter is reopened (like the other 3 settings in the menu currently).


pshriwise (approved):
Simple enough. Thanks @kkiesling!

PR #97, opened 2022-05-31
Adding option to set number of threads when opening the plotter

pshriwise:
Adding an option to set the number of threads OpenMC will use when starting the plotter. Got tired of having to set and reset `OMP_NUM_THREADS` to do this. Used the same CML flags as the `openmc` executable for consistency.


paulromano (approved):
Simple enough -- thanks!

PR #108, opened 2023-01-19
src folder and pyproject packaging

shimwell:
Just seeing if the package could be updated to follow the latest python packaging recommendations https://packaging.python.org/en/latest/tutorials/packaging-projects/

- use of a src folder - helps avoid local import using folder name instead of installed package
- use of pyproject.toml instead of depreciated setup.py
- use of setuptools_scm to get version number

3 benefits of src folder layout are detailed at the bottom of[ this page](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/)


shimwell:
this PR has some major merge conflicts since #110 was merged so I shall close it for now. If there is any interest in moving to a src folder layout for the package, updating the setup to use pyproject.toml or automatic the versioning with setuptools_scm give me a nudge and I can do this PR or aspects of it again

PR #110, opened 2023-02-07
Moving the llc,urc paramaters to the ViewParams class

pshriwise:
I noted today that our VTK export capability was broken b/c the export dialog box couldn't find these when trying to populate initial values. Moving these attributes from the `PlotViewIndependent` class to the `ViewParam` class.


shimwell:
I think @billingsley-john and myself came across this in the York neutronics workshop and couldn't figure out what was going on. Thanks for the fix Patrick

Note to self, merge this will cause conflicts with this other PR #108

kkiesling (approved):
Thanks for fixing this @pshriwise - lgtm

PR #117, opened 2023-04-28
Update github action versions used in python-publish.yml

shimwell:
I've noticed a few deprecation warnings elsewhere and I think v2 of the checkout action now fails due to use of an older version of node js

I also updated the python action to v4 while I was having a look at the file


paulromano (approved):
Thanks @shimwell! :pray:

PR #122, opened 2023-09-12
Include puns colormaps

SteSeg:
Hi! I noticed that the plotter does not allow for perceptually uniform sequential colormaps (e.g. viridis, cividis, plasma etc.). This PR includes them.
It was sufficient to substitute:

`mcolormaps.datad`

with:

`mcolormaps._gen_cmap_registry()`

in [docks.py](https://github.com/openmc-dev/plotter/blob/develop/openmc_plotter/docks.py) and [plotgui.py](https://github.com/openmc-dev/plotter/blob/develop/openmc_plotter/plotgui.py) files.

Picture1

### DIFF
--- a/openmc_plotter/docks.py
+++ b/openmc_plotter/docks.py
@@ -16,6 +16,7 @@ from .custom_widgets import HorizontalLine, Expander
 from .scientific_spin_box import ScientificDoubleSpinBox
 from .plotmodel import (_SCORE_UNITS, _TALLY_VALUES,
                         _REACTION_UNITS, _SPATIAL_FILTERS)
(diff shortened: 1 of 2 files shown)


paulromano (changes requested):
I'm wary of relying on a function that is intended to be private (use of leading underscore). It looks like the same thing can be achieved with:
```Python
import matplotlib.pyplot as plt

plt.colormaps()
```

SteSeg:
Right, I didn't realize that. I switched to those lines now. Thanks!

paulromano (approved):
Thanks @SteSeg!

PR #125, opened 2023-10-07
Instructions for development version install

shimwell:
Just wondering if it is worth adding another install option for people wanting the development version.

The development version occasionally has a few fixes that would help users


shimwell:
Looks like this is no longer needed with a higher frequency of releaes and I guess most developers already know this method of installing anyway

PR #126, opened 2023-10-10
Fix inconsistent model refresh threading

johvincau:
When calling `openmc-plotter` with the `-s` flag and specifying threads, this number would only be remembered for the first time loading the model. Any subsequent reloads called from the GUI would function as if `openmc-plotter` were run with no thread specification, returning to the default value for that machine.

This pull request adds a quick fix to enable any future loads/reloads of the model to use the number of threads as specified by the initial call of `openmc-plotter`.


paulromano (approved):
Thanks for fixing this! Also looks like the reload would have missed out on the `model_path` argument and I see you've included that too :+1:

PR #129, opened 2023-11-10
Pyside6

pshriwise:
This contains what I think is most of an upgrade to PySide6. We'll want to hammer on the features a bit more than I have, but I can explore models and view tally data on both OSX and Linux with these changes.

Aims to resolve #127


pshriwise:
Whoops. Thought #127 was an issue not a PR. I have a couple small changes to add to your PR @eepeterson. I'll add them to your branch in a bit.

PR #133, opened 2024-01-31
added conda install cmd

shimwell:
Just noticed that we don't mention the conda install option.

might be worth adding this to the readme as it sometimes works a bit better than the pip install


paulromano (approved):
Thanks @shimwell!

PR #149, opened 2024-06-06
Casting input value to setData to a CheckState for correct comparison

### LINKED ISSUE #148: Masking check boxes in ColorDialog wont' update
A change in PySide6 for the `Qt.CheckState` type results in an incorrect comparison for the integer values coming into the `QtAbstractTableModel.setData` method. This results in incorrect state for the the checkboxes in the ColorDialog box. The internal masking state does update, but can only go one direction. 

Issue description is here: https://bugreports.qt.io/browse/PYSIDE-1930

pshriwise:
Fixes #148


paulromano (approved):
Thanks for digging into this one!

PR #153, opened 2024-07-24
Mesh annotations

pshriwise:
When building a model with a mesh or applying a mesh, it can be useful to view that mesh on top of the geometry. 

This PR adds mesh annotations, which allow users to view mesh boundaries without having performed a simulation first. This is possible thanks to the nice upgrade on mesh visualization from #134 and applies to all mesh types supported by OpenMC (regular, cylindrical, and unstructured meshes are shown below).

Screenshot 2024-07-24 at 11 49 58 AM

pincell_cm
pincell_rm
pincell_um


openmc_plotter/plotmodel.py
```diff
@@ -746,12 +763,7 @@ def _do_op(array, tally_value, ax=0):
 
         # Get mesh bins from openmc.lib
         mesh_cpp = openmc.lib.meshes[mesh.id]
-        mesh_bins = mesh_cpp.get_plot_bins(
-            origin=origin,
-            width=(view.width, view.height),
-            basis=view.basis,
-            pixels=(view.h_res, view.v_res),
-        )
+        mesh_bins = self.mesh_plot_bins(mesh.id, view)
```
paulromano: This seems to ignore the mesh filter translation now (in my editor, I see that `origin` above is unused)

paulromano (commented):
Thanks for this feature @pshriwise! One question below:

openmc_plotter/plotmodel.py
pshriwise: Thanks for catching that! I made appropriate updates to correct in the latest commit and then verified  that it's still working on the PWR pincell model with a mesh filter translation of (0.2, 0.2, 0.0):

image

paulromano (approved):

PR #160, opened 2024-11-20
Add an CML option to clear the application cache

pshriwise:
A few users have reached out about problems related to crashed caused by the Qt application configuration state, which manifests itself in a few ways. See:

  - #128 
  - https://openmc.discourse.group/t/openmc-plotter-error-object-has-no-attribute-shortcutoverlay/4344
  - https://openmc.discourse.group/t/openmc-plotter-aborts-on-launch-attributeerror-mainwindow-object-has-no-attribute-shortcutoverlay/4158

 Often the answer to these solutions is to remove the plotter configuration file manually (usually in `~/.config/OpenMC`). The `QtSettings` allows one to clear this cache programmatically, so rather than having users taking this step manually this PR provides an option to do so with a `--clear-cache` flag that can be passed via the CML to the plotter application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,19 @@ can be used to generate images from the command line without opening the plotter
 ```console
 $ openmc-plotter -b view1.pltvw view1.pltvw view1.pltvw
 ```
+## Troubleshooting
+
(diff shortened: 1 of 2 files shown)


paulromano (approved):
Thanks @pshriwise!

PR #168, opened 2026-01-19
Implementing blitting for colorbar indicators

paulromano:
I recently tried using the data indicator with a mesh tally displayed in the plotter and found that it was practically unusable because it caused a full redraw of the image each time the mouse moved. This PR changes the indicator updates to use [blitting](https://matplotlib.org/stable/users/explain/animations/blitting.html) so that updating the indicator line can be done quickly without redrawing the full image.

### DIFF
--- a/openmc_plotter/plotgui.py
+++ b/openmc_plotter/plotgui.py
@@ -44,7 +44,7 @@ class PlotImage(FigureCanvas):
         self.x_plot_origin = None
         self.y_plot_origin = None
 
-        self.colorbar = None
+        self.property_colorbar = None
         self.data_indicator = None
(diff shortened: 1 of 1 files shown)


pshriwise (approved):
Looks great! I updated the name of the `PlotImage.colorbar` attribute to something a little more descriptive that was pre-existing (the corresponding colorbar background cache name as well).  Hope that's alright. There was a section of code that was throwing me a bit in `PlotImage._blit_indicator` where a background cache is selected. It helps there in particular I think.

I don't know that I've used the data indicator feature in a while, so thanks for noticing and addressing this!

PR #169, opened 2026-01-23
Add tally colorbar min/max rescaling options

paulromano:
This PR enhances the tally data colorbar controls by replacing the binary "Custom Min/Max" checkbox with a more flexible dropdown offering three scaling options:

- **Full data**: Uses the full data range (previous default behavior)
- **Visible data**: Auto-rescales to only the data currently visible in the plot view
- **Custom**: Allows manual specification of min/max values (previous "Custom Min/Max" behavior)

The new "Visible data" option is particularly useful when zooming in on a region with a narrower data range than the full dataset, allowing for better contrast and visibility without manual adjustment.

### DIFF
--- a/openmc_plotter/docks.py
+++ b/openmc_plotter/docks.py
@@ -773,17 +773,12 @@ class ColorForm(QWidget):
         Selector for colormap
     dataIndicatorCheckBox : QCheckBox
         Inidcates whether or not the data indicator will appear on the colorbar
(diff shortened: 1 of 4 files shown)


pshriwise (approved):
Tried this out and it works quite well. Thanks @paulromano! Pretty handy. I like the layout of the new spinboxes for user defined min/max as well.

PR #170, opened 2026-01-23
Reorganize layout to use tabbed panels instead of docks

paulromano:
This PR restructures the layout of the plotter so that the Geometry, Tallies, and Mesh Annotations docks now appear together in a single dock as two separate tabs. They also share a single green "Apply Changes" button.

Here's an example of what it looks like now:
image


paulromano:
Just added one more update addressing a minor annoyance, which is that the geometry panel was not resizable which prevented the overall GUI window from being resized smaller than a minimum height. I've made all panels resizable so that the window can be resized as desired.

pshriwise (approved):
I really like the new layout. It uses screen real-estate more efficiently and looks cleaner. 

I made a few small tweaks to the mesh tab panel:

  - added a label to indicate the purpose of the tree widget
  - reduced the initial size of the widget and added a splitter to make it resizeable
  - ensured the mesh tree is scrollable

I also removed a couple of references to the old Tally and Mesh docks in the shortcuts overlay.

If this is all agreeable to you @paulromano let me know and we'll get this merged in!

paulromano:
That all looks good -- thanks @pshriwise! The only change I made is to unbold the mesh annotations label since it was the only label that was bold.

PR #171, opened 2026-01-30
Make plot generation asynchronous to keep the UI responsive

paulromano:
This change makes plot generation asynchronous so the UI stays responsive during potentially expensive calls into `openmc.lib`. It introduces a background worker thread for `id_map`/`property_map`, queues plot updates, and updates the view when results are ready. A <s>status‑bar busy indicator</s> translucent overlay on the plot provides feedback while plots are generating. Let me know what you think!

### DIFF
--- a/openmc_plotter/main_window.py
+++ b/openmc_plotter/main_window.py
@@ -56,8 +56,13 @@ class MainWindow(QMainWindow):
         self.model_path = Path(model_path)
         self.threads = threads
         self.default_res = resolution
+        self.model = None
         self.plot_manager = None
-        self.busyIndicator = None
+        self.busyIndicator = QProgressBar()
+        self.busyIndicator.setRange(0, 0)
+        self.busyIndicator.setMaximumWidth(self.font_metric.averageCharWidth() * 12)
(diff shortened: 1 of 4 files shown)


pshriwise:
This is really nice. Being able to interact with the app window while plots are being generated is a huge plus. 

I like the status indicator and status bar text idea, though the status bar indicator was a little subtle to me, particularly on a wider monitor. I tried combining the two into an overlay in the plot image area. What do you think?

[Screencast from 2026-02-05 15-23-34.webm](https://github.com/user-attachments/assets/d0e17896-766c-4324-a1d4-caf81e5b28fc)

paulromano:
@pshriwise Interesting, can you throw that on another branch for me to play around with? If I like it, I'll merge it into the branch here

pshriwise:
Here ya go! https://github.com/pshriwise/plotter/tree/update-overlay

paulromano:
I do indeed like your version better and merged it in; thanks! I made a few more small updates -- let me know what you think now.

pshriwise (approved):
Thanks @paulromano!

PR #172, opened 2026-02-14
Use new `slice_data` function from OpenMC

paulromano:
This PR replaces the calls to `id_map` and `property_map` with the single `slice_data` function introduced in openmc-dev/openmc#3806. The filter binning capability is currently only utilized for `MeshMaterialFilter` but with some more refactoring, we should be able to use it for all spatial tally filters, which will remove a lot of code in `PlotModel`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,23 +24,26 @@ jobs:
     env:
       DISPLAY: ':99.0'
     steps:
-      -
-        name: Apt dependencies
+      - name: Apt dependencies
         shell: bash
         run: |
           apt update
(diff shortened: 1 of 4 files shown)


pshriwise:
@paulromano do you feel this is ready to mark for review?

paulromano:
@pshriwise Yes, I had marked it as draft since it relies on functionality that hasn't been merged to OpenMC yet. However, now that we've gone through one round of reviews there, I'd say this is ready for review (understanding that we shouldn't merge until openmc-dev/openmc#3806 goes in)

pshriwise (approved):
Looks good from here @paulromano! Thank you for updating the API. I noted a significant improvement in plotting speed as promised :)

PR #173, opened 2026-02-15
Add dialogue for adjusting font size and type

pshriwise:
I've often wanted to use the images produced by the plotter for publication but find that the labels in the image are too small to be legible. This provides an option to adjust the axis and colorbar tick mark and label sizes. It also includes the ability to change the font type for the set of fonts available in the user's `matplotlib` installation.


shimwell:
interesting, would this mean we can avoid using matplotlib.mpl.rcParams which I have a the top of a lot of scripts making the plots font size nice for publications or bigger for presentations

PR #174, opened 2026-02-16
Render widget

pshriwise:
This adds a rendering widget to the plotter based on the raytracing plot capability in OpenMC. A big thanks to @gridley for submitting the enabling-feature in OpenMC originally!

A rendering window can be opened using a "Render" button in the Geometry panel. An orbit camera implementation has been provided.

Colors of domains in the render widget and main window are synchronized and can be changed in either place. Live camera parameters are provided in the widget to provide a frame of reference for slice plots. Domain visibility can be toggled and the light source position can be modified independently of the camera position if desired.

There are a litany of new features that come to mind with this tool, but I don't want to submit something overwhelming. 

A good amount of this was generated with Codex, so I'm marking it as a draft while I review it and get feedback from others in the meantime.

---

Demos:

### Pincell

[Screencast from 2026-02-16 12-07-10.webm](https://github.com/user-attachments/assets/2ce5a1d0-3d68-4ddb-a898-0052ed2e3a15)

### TRISO Assembly

I'm really pleased that we can view these models interactively. I was surprised at this.

[Screencast from 2026-02-16 12-09-31.webm](https://github.com/user-attachments/assets/ed98cb08-5f11-4f3f-ba5e-4fb618f31d53)

### DAGMC Model (Utah Teapot)

This one isn't quite as interactive as I'd like. I have an idea or two as to why, but it's also possible we might simply be stuck with that for DAGMC models at present.

[Screencast from 2026-02-16 12-23-16.webm](https://github.com/user-attachments/assets/669232e9-8bb1-4a15-8edc-8eaec31fef2f)

### DIFF
--- a/openmc_plotter/docks.py
+++ b/openmc_plotter/docks.py
@@ -184,8 +184,8 @@ class GeometryPanel(PlotterPanel):
         self.panelLayout.addWidget(self.resGroupBox)
         self.panelLayout.addWidget(HorizontalLine())
         self.panelLayout.addWidget(self.zoomWidget)
-        self.panelLayout.addStretch()
         self.panelLayout.addWidget(self.renderButton)
+        self.panelLayout.addStretch()
 
         self.setLayout(self.panelLayout)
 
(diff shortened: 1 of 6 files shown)


gridley:
🤯 awesome!!

PR #175, opened 2026-03-06
Make material info context window scrollable

pshriwise:
For materials with enough nuclides, the list can run beyond the height of a user's screen. At this point, some of the nuclides are not able to be viewed. These changes make that context window scroll-able, so the entire material composition can be viewed.


paulromano (approved):
Nice, just tried it out on a model with some complex materials and it worked nicely :+1:

PR #176, opened 2026-03-06
Copy image to clipboard

pshriwise:
In most cases these days it's possible to right click and copy an image directly out of a browser, chat, etc. 

These changes make that possible and open them to see plotter as well.


paulromano (approved):
I like it :+1:

PR #177, opened 2026-03-08
Allow opening statepoint files when one is already open

pshriwise:
Another QOL upgrade here. It's been bothering me for a while that you can't more cleanly open a new statepoint file with one already open. Feels clunky for it to fail and force the user to close the existing file manually. This takes care of that by providing a notice if one is open and optionally opening the file open dialog anyway after confirmation.


pshriwise:
The test code is fully codex-generated. It's a lot of line changes and while I'd like to have more testing here it's not the type of testing I'm primarily concerned with. Happy to remove it if we think that's best.

paulromano (approved):
Sorry for delay on this; thanks!

PR #179, opened 2026-07-02
Overlapping Visualization for Plotter

viktormai:
This PR includes the plotter-side code corresponding to #3969 in the OpenMC main branch, allowing the plotter to use the information received from the C API via `openmc.lib.slice_data_overlap_info()`. A simple input file will be included for visualization purposes of overlapping regions with the plotter. 
[overlap_test.py](https://github.com/user-attachments/files/29607103/overlap_test.py)

**Changes**

- `getIDinfo`: Read hover IDs directly from cell_ids/mat_ids (i.e. raw geom_data) instead of the processed self.ids. This preserves the encoded overlap index (e.g. -4, -5) which is needed to look up the correct entry in overlap_map.
- `mouseMoveEvent:` Checks `elif id < _OVERLAP`, which decodes the overlap index via `_OVERLAP - id - 1` and looks it up in `overlap_map` to display universe and cell info. Cell names are shown if available, falling back to IDs.
- `contextMenuEvent`: Change `id not in (_NOT_FOUND, _OVERLAP)` to `id >= _OVERLAP` so that overlap pixels (which now carry values more negative than -3) are correctly excluded from the domain color/mask/highlight context menu.
- `overlap_map`: Initialized to {} at model construction and populated after every `makePlot` call by reading `slice_data_overlap_info()`. 
- `self.ids` is now a copy of cell_ids/mat_ids rather than a direct reference. This is necessary because the normalization step mutates the array in place — without the copy it would corrupt `geom_data` itself, breaking the raw index lookups in `getIDinfo`.
- Normalization: `self.ids[self.ids < _OVERLAP] = _OVERLAP` clamps all encoded overlap indices down to -3 in `self.ids`. This is required because `self.ids` is used for image construction via `domain[id].color`, and `DomainViewDict` only has entries for real domain IDs plus the sentinel values `_OVERLAP`, `_NOT_FOUND`, and `_VOID_REGION` — passing -4, -5, etc. would raise a `KeyError`. The raw values are still accessible via cell_ids/mat_ids for hover lookup.

### DIFF
--- a/openmc_plotter/plotgui.py
+++ b/openmc_plotter/plotgui.py
@@ -309,7 +309,10 @@ class PlotImage(FigureCanvas):
         # check that the position is in the axes view
         if 0 <= yPos < self.model.currentView.v_res \
            and 0 <= xPos and xPos < self.model.currentView.h_res:
-            id = self.model.ids[yPos, xPos]
+            if self.model.currentView.colorby == 'cell':
+                id = int(self.model.cell_ids[yPos, xPos])
+            else:
+                id = int(self.model.mat_ids[yPos, xPos])            
             instance = self.model.instances[yPos, xPos]
             temp = "{:g}".format(self.model.property_data[yPos, xPos, 0])
             density = "{:g}".format(self.model.property_data[yPos, xPos, 1])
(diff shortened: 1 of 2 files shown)


paulromano:
@viktormai I just updated your branch to fix an exception I was running into when switching the color-by from material to cell and also match the updated interface as proposed in openmc-dev/openmc#4002. Once that's merged, this should be good to go!

paulromano (approved):

